PR #5, opened 2023-11-02
feat: disable candidates

baszalmstra:
Adds the ability to disable specific candidates within the solver. This can be useful for error reporting. 

Previously one would filter out incompatible candidates before passing them to the solver. This has the downside that if the problem is unsolvable the filtering is not transparent to the user. With this change the solver receives the candidates and immediately deselects them. However, they are still present in the error output with a custom message:

```text
The following packages are incompatible
|-- a * cannot be installed because there are no viable options:
    |-- a 1 is disabled because it is externally disabled
```

The `it is externally disabled` in this example was injected into the solver.

I think in a next iteration it would be nice to merge some of these problems together to get a more readable output. This is currently a bit verbose if there are many disabled versions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,11 +107,11 @@ pub struct Candidates {
     /// dependencies can easily be provided one should provide them up-front.
     pub hint_dependencies_available: Vec<SolvableId>,
 
-    /// A list of solvables that are available but have been disabled for reasons outside of the
-    /// solver. For example, a package might be disabled because it is not compatible with the
+    /// A list of solvables that are available but have been excluded from the solver. For example,
+    /// a package might be excluded from the solver because it is not compatible with the
     /// runtime. The solver will not consider these solvables when forming a solution but will use
     /// them in the error message if no solution could be found.
-    pub disabled: Vec<(SolvableId, StringId)>,
+    pub excluded: Vec<(SolvableId, StringId)>,
 }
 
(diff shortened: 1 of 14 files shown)


src/lib.rs
```diff
@@ -104,6 +106,12 @@ pub struct Candidates {
     /// solver doesnt actually need this information to form a solution. In general though, if the
     /// dependencies can easily be provided one should provide them up-front.
     pub hint_dependencies_available: Vec<SolvableId>,
+
+    /// A list of solvables that are available but have been disabled for reasons outside of the
+    /// solver. For example, a package might be disabled because it is not compatible with the the
```
wolfv, line 111: Duplicate `the` at the end of the comment

wolfv (approved):
Just a little typo. Looks great to me, excited for this!

src/solver/clause.rs
```diff
@@ -265,6 +279,11 @@ impl ClauseState {
         Self::from_kind_and_initial_watches(kind, watched_literals)
     }
 
+    pub fn disable(candidate: SolvableId, reason: StringId) -> Self {
+        let (kind, watched_literals) = Clause::exclude(candidate, reason);
+        Self::from_kind_and_initial_watches(kind, watched_literals)
+    }
+
```
tdejager, line 286: Should this also be exclude?

PR #11, opened 2024-01-11
add version match options

wolfv:
I tried something to be able to pass in some additional options when trying to match versions (e.g. wether pre-releases are allowed globally or not).


baszalmstra:
Couldnt you achieve the same by storing an Rc to the options in the versionset itself?

PR #15, opened 2024-01-24
fix: incremental solving (rip issue #75)

### LINKED ISSUE #13: resolver issue where unreachable code is reached
We hit the following – supposedly unreachable code – in certain cases:

```
thread 'main' panicked at 'internal error: entered unreachable code: when we get here it means that all candidates have been assigned false. This should not be able to happen at this point because during propagation the solvable should have been assigned false as well.', /Users/tdejager/.cargo/registry/src/index.crates.io-6f17d22bba15001f/resolvo-0.2.0/src/solver/mod.rs:584:53
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

A test has been added to resolvo that recreates the issue #10.

aochagavia:
The original bug was caused by adding a conflicting `requires` clause, yet failing to detect and handle the conflict. This commit introduces a clearer separation between: adding new clauses, detecting conflicts, and handling said conflicts.

Closes #13 

Closes https://github.com/prefix-dev/rip/issues/75

### DIFF
--- a/src/solver/mod.rs
+++ b/src/solver/mod.rs
@@ -36,6 +36,8 @@ pub struct Solver<VS: VersionSet, N: PackageName, D: DependencyProvider<VS, N>> 
     requires_clauses: Vec<(SolvableId, VersionSetId, ClauseId)>,
     watches: WatchMap,
 
+    negative_assertions: Vec<(SolvableId, ClauseId)>,
+
     learnt_clauses: Arena<LearntClauseId, Vec<Literal>>,
     learnt_why: Mapping<LearntClauseId, Vec<ClauseId>>,
     learnt_clause_ids: Vec<ClauseId>,
@@ -57,6 +59,7 @@ impl<VS: VersionSet, N: PackageName, D: DependencyProvider<VS, N>> Solver<VS, N,
             clauses: Arena::new(),
             requires_clauses: Default::default(),
             watches: WatchMap::new(),
+            negative_assertions: Vec::new(),
             learnt_clauses: Arena::new(),
             learnt_why: Mapping::new(),
             learnt_clause_ids: Vec::new(),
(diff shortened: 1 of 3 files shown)


src/solver/mod.rs
```diff
@@ -732,8 +684,25 @@ impl<VS: VersionSet, N: PackageName + Display, D: DependencyProvider<VS, N>> Sol
     /// way, the clause can check whether the literal that is using the solvable has become false, in
     /// which case it picks a new solvable to watch (if available) or triggers an assignment.
     fn propagate(&mut self, level: u32) -> Result<(), (SolvableId, bool, ClauseId)> {
-        // Learnt assertions (assertions are clauses that consist of a single literal, and therefore
-        // do not have watches)
+        // Negative assertions derived from other rules (assertions are clauses that consist of a
+        // single literal, and therefore do not have watches)
+        for &(solvable_id, clause_id) in &self.negative_assertions {
```
baszalmstra, line 689: Propagation is done a lot, this now runs for all `negative_assertions` every time `propagate` is called. Would it make sense to only do this for a subset of the negative assertions?

wolfv: Should we rerun the benchmarks in `rattler`?

aochagavia: I think running the benchmarks makes sense. It's been a long time since I wrote this code, but I was under the impression that this part of propagation is not run _that_ often (only after an arbitrary decision has been made). Also, it's not clear to me how to select the right subset of negative assertions without introducing complexity. With a bit of luck, the perf impact will be negligible and we can get away with it.

wolfv (approved):

src/solver/mod.rs
aochagavia: (For future reference: the benchmarks showed no significant difference / a slight improvement)

aochagavia:
@sumanth-manchala thanks for creating the test that allowed me to develop this fix! I credited you in my [last blog article](https://ochagavia.nl/blog/chasing-a-bug-in-a-sat-solver/) :)

sumanth-manchala:
That was really unexpected!!! Thanks, means a ton❤️ and that int test was my first open source contribution!!

aochagavia:
Great to hear :)

For me, this is the 10th year since my first contribution. You can have a look at [my first PR](https://github.com/rust-lang/rust/pull/12871) and you will have a good laugh, because I didn't know git that well back then 😅

PR #17, opened 2024-01-24
Allow returning "unknown dependencies" in `get_dependencies`

### LINKED ISSUE #14: Exclude candidates in `get_dependencies` of the DependencyProvider
# Description

It would be a really nice feature if we have a way to exclude candidates when getting their dependencies. 
With pypi packages for example it's possible that you are able to get the candidates for a Solvable but that it is impossible to get the dependencies for this candidate. 

With pypi packages, specifically in the case of source distributions, you might not be able to get the metadata because you have to build the source, which might fail. It would be nice to be able to exclude this candidate at this point.

However, we would need to think how to best handle this, as backtracking might not be interesting. Take the case where you want to have `pygame` as a dependency for example, this depends on `sdl2`, should this source build fail because `sdl2` is missing. Backtracking will take long (because it has to build everything) and would not produce any good results. I think aborting the solve in this case would be best. However, there might be situations where an older build would work, idk, maybe backtracking does make sense.

# Prior art

When looking at pubgrub, specifically: https://github.com/pubgrub-rs/pubgrub/blob/717289be5722dd5caaa0d1f4ed13047d11a7f7fd/src/solver.rs#L141C8-L153C21

It seems there are 2 options for `get_dependencies`, there is a `Result` that makes the solver abort the current resolution. But also a `Dependencies::Unknown` enum that adds an `incompatibility` to the solver.

aochagavia:
Note: this assumes #15 and #16 are merged.

Question: should we make it possible for the provider to tell the reason why the dependencies are unknown? For instance, a string where we can put the contents of [this panic](https://github.com/prefix-dev/rip/blob/25e2344e62804dab039a1c806b7489c3779f3e47/crates/rattler_installs_packages/src/resolve/dependency_provider.rs#L475-L478)

Closes #14

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,7 +89,7 @@ pub struct Candidates {
     /// no solution could be found otherwise.
     ///
     /// The same behavior can be achieved by sorting this candidate to the top using the
-    /// [`DependencyProvider::sort_candidates`] function but using this method providers better
+    /// [`DependencyProvider::sort_candidates`] function but using this method provides better
     /// error messages to the user.
     pub favored: Option<SolvableId>,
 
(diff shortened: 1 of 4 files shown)


baszalmstra:
You are on 🔥! And yes to the question! Because that would be ideal to show in a unsat report

aochagavia:
All right, just pushed the changes.

And here's [a diff](https://github.com/aochagavia/rip/commit/8ca8c43e26903cfe739c48c4f71d6e4bcb15807b#diff-d0966c32b9e9e133844c038b8114768a48d80dd9e9bfe54a3654f408e5d5c5b6) showing how it looks like in rip.

Eh2406:
BTW, I am hoping to add a user customizable error string for unknown dependencies in pubgrub. So I think it is a good idea here too!

wolfv:
Nice work @aochagavia! :) Let's get your PRs reviewed and merged tomorrow 🚀

baszalmstra (approved):

tdejager (approved):

PR #28, opened 2024-01-31
wip: experimental support for version ranges in unsat errors

aochagavia:
I won't have time to finish up this PR properly, but I'm leaving it here because at least part of it might be interesting. The second commit could probably be cherry-picked and merged without too much additional work.

### Experimental support for version ranges

@wolfv was enthusiastic about the idea, but we didn't really reach a consensus about how the feature would exactly work. The first commit contains a working implementation that you could tweak further. Check out the insta snapshots of the commit to see how it affects error messages.

### Proper unicode characters in error messages

This means we stop using `|--` and start using `├─`, and similar. There is some complexity associated to this change, because now we need to do some tracking to know which character we need (`├─`, `└─` or `│`). The implementation is in the second commit, but there are a few TODOs to clean up the code. Check out the insta snapshots of the commit to see how it affects error messages.

### DIFF
--- a/tests/snapshots/solver__merge_installable_non_continuous_range.snap
+++ b/tests/snapshots/solver__merge_installable_non_continuous_range.snap
@@ -1,0 +1,10 @@
+---
+source: tests/solver.rs
+expression: "solve_snapshot(provider, &[\"a 1\", \"a 2..20\"])"
+---
+The following packages are incompatible
+|-- a >=2, <20 can be installed with any of the following options:
+    |-- a 19 | >=16, <=17 | >=9, <=14 | >=2, <=7
+|-- a >=1, <2 cannot be installed because there are no viable options:
+    |-- a 1, which conflicts with the versions reported above.
+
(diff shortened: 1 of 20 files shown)


tests/snapshots/solver__merge_installable_non_continuous_range.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: tests/solver.rs
+expression: "solve_snapshot(provider, &[\"a 1\", \"a 2..20\"])"
+---
+The following packages are incompatible
+├─ a >=2, <20 can be installed with any of the following options:
+│  └─ a 19 | >=16, <=17 | >=9, <=14 | >=2, <=7
```
wolfv, line 7: I was told that `pubgrub` has a implementation of a `simplify` algorithm for ranges like this (it would go and merge all the ranges that overlap so that the result would be the same range but expressed with fewer terms). I think in this case, it would be nice if it was simplified to `>=2,<20` - although that _is_ in fact what the "MatchSpec" above says ... so this thought might be futile anyways.

baszalmstra:
Closing this PR for now. We don't see an avenue forward for this as of yet.

PR #30, opened 2024-02-08
feat: improved indentation in error messages

baszalmstra:
I extracted the new indentation code from #28 

**Old**
```
The following packages are incompatible
|-- a * cannot be installed because there are no viable options:
    |-- a 2 would require
        |-- b *, which cannot be installed because there are no viable options:
            |-- b 1 is excluded because it is externally excluded
    |-- a 1 would require
        |-- c *, which cannot be installed because there are no viable options:
            |-- c 1 is excluded because it is externally excluded
```

**New**
```
The following packages are incompatible
└─ a * cannot be installed because there are no viable options:
   ├─ a 2 would require
   │  └─ b *, which cannot be installed because there are no viable options:
   │     └─ b 1 is excluded because it is externally excluded
   └─ a 1 would require
      └─ c *, which cannot be installed because there are no viable options:
         └─ c 1 is excluded because it is externally excluded
```

### DIFF
--- a/src/problem.rs
+++ b/src/problem.rs
@@ -508,6 +508,114 @@ impl ProblemGraph {
     }
 }
 
+#[derive(Copy, Clone, PartialEq, Eq)]
+enum ChildOrder {
+    HasRemainingSiblings,
+    Last,
+}
+
+struct Indenter {
+    levels: Vec<ChildOrder>,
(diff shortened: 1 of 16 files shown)


wolfv (approved):
Looking gooood!

PR #35, opened 2024-05-28
perf: small memory performance optimizations

baszalmstra:
* Reduce the size of some datastructures
* Use `ahash` for hash tables

Provides a 1-5% performance increase.


wolfv:
Should we merge this one?

PR #36, opened 2024-05-30
refactor: reduce memory usage

baszalmstra:
Reduce the total memory used by the solver by grouping similar datatypes together. There is still some unsafe code in here, which I want to try and clean up a little more.


baszalmstra:
With #66 the memory usage has already decreased. Closing this for now.

PR #44, opened 2024-06-10
feat: add `DependencySnapshot`

baszalmstra:
This PR adds a `DependencySnapshot` type. This type can be created from anything that implements `DependencyProvider`. The snapshot can be serialized and later be used as a `DependencyProvider`. This allows capturing ecosystem-specific implementations in a single file format which makes it great for running extensive benchmarks for different ecosystems.

Creating a snapshot requires an ecosystem implementation. An example can be found in https://github.com/mamba-org/rattler/pull/741.

I also included a little tool called `solve-snapshot` that will solve for all package names found in a single snapshot and output timing information.


wolfv (approved):

Eh2406:
The similar concept we have in pubgrub has been working really well for us! I hope it works well for you as well.

PR #48, opened 2024-06-15
fix: display_merged_solvables to display merged solvables correctly

jerily:
minor fix in display_merged_solvables so that it displays them correctly. without this fix a merged solvable is displayed as "a21" as opposed to "a 2 | 1".


baszalmstra:
Thanks! Much appreciated!

PR #50, opened 2024-06-24
fix: Add a virtual destructor to `DependencyProvider`

jjerphan:
When subclassing `DependencyProvider`, I currently get this warning:

```
warning: base class 'struct resolvo::DependencyProvider' has accessible non-virtual destructor [-Wnon-virtual-dtor]
   53 | struct PackageDatabase : public DependencyProvider {
      |        ^~~~~~~~~~~~~~~
```

I think any base class which has at least one virtual function, must have virtual destructor to respect class polymorphism.


baszalmstra:
Can you explain why that is needed? The code that interfaces with the provider doesnt destruct it, does it?

baszalmstra:
Nvm, did my research! 🧐 LGTM!

PR #54, opened 2024-07-15
feat(solver)!: Solve for optional solvables in addition to the root solvable

eviltak:
Adds a new `Solver::solve_with_additional` method, that solves for the provided `root_requirements` and `root_constraints`, along with a set of additional (optional) solvables. Unless the corresponding package has been requested by a version set in another solvable's clauses, each additional solvable is _not_ subject to the package-level clauses introduced in `DependencyProvider::get_candidates` since the solvables have been requested specifically (and not through a version set).

This feature is useful for situations where you want to ensure that particular solvables (for example, previously installed solvables) are compatible with the solution to the root requirements and constraints.

The implementation is facilitated by making the `run_sat` method solve for arbitrary solvables at arbitrary levels, not just the root solvable. This allows the solver to solve for the additional solvables after solving for the root solvable while taking into account the clauses and decisions introduced in the solving process.

# Breaking changes
- Renames the `problem` module and its container structs to `conflict`, since they are used to inspect an unsatisfiable conflict
- Uses the newly added `solver::Problem` struct as an input for the `Solver::solve` method to allow for flexibility in solver input arguments

Existing users will have to change the code that calls `Solver::solve` to pass in an instance of the `Problem` struct rather than passing in the `root_requirements` and `root_constraints` directly as arguments to the method.

### DIFF
--- a/src/solver/mod.rs
+++ b/src/solver/mod.rs
@@ -182,7 +182,7 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
         assert_eq!(root_clause, ClauseId::install_root());
 
         // Run SAT
-        self.run_sat()?;
+        self.run_sat(InternalSolvableId::root())?;
 
         let steps: Vec<SolvableId> = self
             .decision_tracker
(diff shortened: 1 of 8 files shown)


baszalmstra:
Sorry, Im a bit busy with other things atm! Will get back to this PR as soon as I can.

src/solver/mod.rs
```diff
@@ -167,6 +172,78 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
...
+    ///
+    /// Returns [`UnsolvableOrCancelled::Unsolvable`] only if the _root_ requirements and
+    /// constraints could not be solved for; if an additional solvable is unsolvable, it is
+    /// simply not included in the solution.
+    pub fn solve_with_additional(
+        &mut self,
+        root_requirements: Vec<VersionSetId>,
+        root_constraints: Vec<VersionSetId>,
+        additional_solvables: impl IntoIterator<Item = SolvableId>,
+    ) -> Result<impl Iterator<Item = SolvableId> + '_, UnsolvableOrCancelled> {
```
baszalmstra, line 216: Instead of having a separate function, I propose we break the API and introduce a struct over multiple arguments. Otherwise, I'm a bit worried that in the future we will have more options and have to add more "overloads". 

Ideally, this struct implements the `Default` trait. So you can write something like:

```rust
solver.solve(SolveGoal { root_requirements: vec![..], ..SolveGoal::default() })
```

We can also rename `root_requirements` to `requirements`, the same for `constraints`. I would also like to rename `additional_solvables` to `soft_requirements`. In the future, we will have to break the API again when we change from solvables to requirements but I think the name `requirements` conveys better what the intent is. In the doc comment for that field we should be very clear that we know currently it only accepts solvables and we will most likely change that in the future.

Would you be able to make these changes yourself?

src/solver/mod.rs
```diff
@@ -579,45 +629,55 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
...
     ///
     /// The solver loop can be found in [`Solver::resolve_dependencies`].
-    fn run_sat(&mut self) -> Result<(), UnsolvableOrCancelled> {
-        assert!(self.decision_tracker.is_empty());
-        let mut level = 0;
+    ///
+    /// Returns `Ok(true)` if a solution was found for `solvable`. If a solution was not
+    /// found, returns `Ok(false)` if some decisions have already been made by the solver
+    /// (i.e. the decision tracker stack is not empty). Otherwise, returns `Err` on no solution.
+    fn run_sat(&mut self, solvable: InternalSolvableId) -> Result<bool, UnsolvableOrCancelled> {
```
baszalmstra, line 636: I think the return type doesnt quite work. I would expect the `Cancelled` return value to still be returned when a cancellation is triggered even for the additional solvables. Why not just always return `UnsolvableOrCancelled` and catch the unsolvable error in `solve_with_additional`.

eviltak: `Err(Cancelled)` should still be returned - the only change is returning `Ok(false)` instead of `Err(Unsolvable)` for unsolvable additional solvables. I did not go for the approach you recommended because it would lead to calling `analyze_unsolvable` being unnecessarily called (though I suppose the compiler may be able to optimize it away). `analyze_unsolvable` also currently seems to assume (via a `debug_assert_eq` statement ) that the highest level of decisions made is 1, and I didn't want to break the assumption (the level would be higher than 1 for unsolvable additional solvables).

src/solver/mod.rs
eviltak: I should be able to, yes. Would it make sense to have `SolveGoal` (or the equivalent) to have private fields and provide constructor methods/builders to make future additions non-breaking changes, or should we just keep all fields public and allow API breakages when adding a new argument (with the assumption that people using `..SolveGoal::default()` will be unaffected?

baszalmstra: Let's go with a struct for now because that makes the C++ implementation simpler.

src/solver/mod.rs
baszalmstra: 👍 Maybe clarify that in the comment.

eviltak: Updated the doc comment to specify that it still returns `Err(Cancelled)`. Did you also want me to include an explanation as to why I don't return `Unsolvable` for additional solvables?

baszalmstra: Nope this is fine! Thank you :)

src/solver/mod.rs
eviltak: I think `Problem` would be a good name for this input struct (since it defines the problem the solver is trying to solve for). Would it make sense to rename the existing `Problem` structs in the `solver::problem` submodule to something a bit more relevant, such as `Conflict`, since they describe the nature of an unsatisfiable conflict, rather than the problem being solved for itself?

baszalmstra: Sounds good!

baszalmstra (approved):
Merging your other PR has yielded a number of conflicts here. Would you be able to resolve those? Ill merge right after! 👍

eviltak:
I'll do this in the next few hours, so it should be ready for you to merge tomorrow. Thanks for your help!

eviltak:
Rebased onto `main` with all conflicts fixed.

baszalmstra:
Thank you!

PR #55, opened 2024-07-17
chore: Add more tracing

jjerphan:


baszalmstra (changes requested):
Thanks for this! 

You changed a lot of places to display ids instead of the rhings the ids represent. I can understand that this makes debugging some things easier but it also makes following the general flow of the algorithm much harder. Could we maybe show both? The id and the representation of the id? 

You inplemented Display in a few places in a way that almost exactly mimics the (derived) Debug trait. Maybe you can use rhat instead?

I think the levels are sometimes quite high. Perhaps consider downgrading some info and debug logs to trace level instead?

src/solver/cache.rs
```diff
@@ -172,13 +172,24 @@ impl<D: DependencyProvider> SolverCache<D> {
             Some(candidates) => Ok(candidates),
             None => {
                 let package_name = self.provider.version_set_name(version_set_id);
+
+                tracing::debug!(
```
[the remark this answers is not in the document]
jjerphan: Yes, let me change all the new `info` and `debug` to `trace`.

src/solver/mod.rs
```diff
@@ -34,6 +34,12 @@ struct AddClauseOutput {
     clauses_to_watch: Vec<ClauseId>,
 }
 
+impl Display for AddClauseOutput {
```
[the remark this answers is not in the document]
jjerphan: I had to implement it, but I guess I should be able to change part of the implementation so that the `Debug` trait is used.

src/internal/id.rs
```diff
@@ -60,6 +60,12 @@ impl ArenaId for VersionSetId {
 #[cfg_attr(feature = "serde", serde(transparent))]
 pub struct SolvableId(pub u32);
 
+impl SolvableId {
+    pub(crate) const fn as_internal(self) -> InternalSolvableId {
+        InternalSolvableId(self.0 + 1)
+    }
+}
```
baszalmstra, line 67: You can just use `InternalSolvableId::from(id)` instead!

src/solver/cache.rs
```diff
@@ -171,14 +171,26 @@ impl<D: DependencyProvider> SolverCache<D> {
...
             Some(candidates) => Ok(candidates),
             None => {
-                let package_name = self.provider.version_set_name(version_set_id);
-                let candidates = self.get_or_cache_candidates(package_name).await?;
+                let package_name_id = self.provider.version_set_name(version_set_id);
+
+                tracing::trace!(
+                    "Getting matching candidates for package: {:?}",
+                    self.provider.display_name(package_name_id).to_string()
+                );
```
baszalmstra, line 179: `display_name` returns something that implements `Display`. You can then use `{}` to format it directly without having to convert to a string first:

```suggestion
                tracing::trace!(
                    "Getting matching candidates for package: {}",
                    self.provider.display_name(package_name_id)
                );
```

src/solver/mod.rs
```diff
@@ -1002,6 +1067,11 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
         while let Some(decision) = self.decision_tracker.next_unpropagated() {
             let pkg = decision.solvable_id;
 
+            tracing::trace!(
+                "Exploring watched solvables for {}",
+                pkg.display(self.provider())
+            );
```
baszalmstra, line 1073: I ran a little test and this popped-up quite a lot, is this useful to see in the logs?

src/solver/mod.rs
```diff
@@ -939,25 +987,42 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
     /// solvable has become false, in which case it picks a new solvable to
     /// watch (if available) or triggers an assignment.
     fn propagate(&mut self, level: u32) -> Result<(), PropagationError> {
+        tracing::trace!("Propagate");
```
baszalmstra, line 990: Is this useful? It always follows an "install". Perhaps add the level at least?

src/solver/mod.rs
```diff
@@ -939,25 +987,42 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
...
             return Err(PropagationError::Cancelled(value));
         };
 
         // Negative assertions derived from other rules (assertions are clauses that
         // consist of a single literal, and therefore do not have watches)
         for &(solvable_id, clause_id) in &self.negative_assertions {
+            tracing::trace!(
+                "Negative assertions derived from other rules: {} = false",
+                solvable_id.display(self.provider())
+            );
```
baszalmstra, line 1002: This is added very often to the logs because its visited on every propagation.

src/solver/mod.rs
```diff
@@ -939,25 +987,42 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
...
             let value = false;
             let decided = self
                 .decision_tracker
                 .try_add_decision(Decision::new(solvable_id, value, clause_id), level)
                 .map_err(|_| PropagationError::Conflict(solvable_id, value, clause_id))?;
 
+            tracing::trace!(
+                "Negative assertions derived from other rules: Decided: {}",
+                decided
+            );
```
baszalmstra, line 1013: Same here. Only the one below seems interesting.

baszalmstra (approved):

PR #56, opened 2024-07-29
feat!: Version set unions as solvable requirements

eviltak:
This PR introduces the concept of _version set unions_ and allows them to be used to specify solvable requirements that can be fulfilled by more than one version set (belonging to more than one package).

The change is facilitated by the introduction of a `Requirement` type, which is used instead of `VersionSetId`s to specify a required dependency of a solvable. A requirement can either be comprised of a single version set (a `VersionSetId`), or multiple version sets (a `VersionSetUnionId`). This allows existing code bases to continue using a single version set (`VersionSetId`) as a requirement specification without ever having to touch version set unions (`VersionSetUnionId`).

# Breaking changes
This PR introduces _breaking_ changes - the main ones being the following:

- Added the `Interner::version_sets_in_union` trait method 
- Changed the type of the `requirements` field of  `KnownDependencies` from `Vec<VersionSetId>` to `Vec<Requirement>`.
- Renamed the `requirements` field of `DependencySnapshot` to `version_sets`
- Added the `version_set_unions` field to `DependencySnapshot` (only affects those who are constructing `DependencySnapshot` instances manually)

## API migration
Migrating existing depending projects that do not rely on version set unions to the new API is quite straightforward.
- Add a placeholder implementation for `version_sets_in_union` in impls of `Interner`. It is perfectly fine for the implementation to be `unimplemented!`, since if version set unions are unused this method will never be called.
- Modify implementations of `DependencyProvider::get_dependencies` to convert the `VersionSetId`s used in the `requirements` field of  `KnownDependencies` to `Requirement`s. Since `Requirement` implements `From<VersionSetId>`, this is as simple as calling `From::from` or `Into::into` on each `VersionSetId`.
-  For those using `DependencySnapshot`, the `requirements` field has been renamed to `version_sets`; all other functionality remains the same.

### DIFF
--- a/src/utils/pool.rs
+++ b/src/utils/pool.rs
@@ -6,7 +6,8 @@ use std::{
 use crate::internal::{
     arena::Arena,
     frozen_copy_map::FrozenCopyMap,
-    id::{NameId, SolvableId, StringId, VersionSetId},
+    id::{NameId, SolvableId, StringId, VersionSetId, VersionSetUnionId},
+    small_vec::SmallVec,
 };
 
 /// A solvable represents a single candidate of a package.
@@ -47,6 +48,8 @@ pub struct Pool<VS: VersionSet, N: PackageName = String> {
 
     /// Map from version set to the id of their interned counterpart
     version_set_to_id: FrozenCopyMap<(NameId, VS), VersionSetId, ahash::RandomState>,
+
+    version_set_unions: Arena<VersionSetUnionId, SmallVec<VersionSetId>>,
 }
 
 impl<VS: VersionSet, N: PackageName> Default for Pool<VS, N> {
@@ -61,6 +64,7 @@ impl<VS: VersionSet, N: PackageName> Default for Pool<VS, N> {
             string_to_ids: Default::default(),
             version_set_to_id: Default::default(),
             version_sets: Arena::new(),
+            version_set_unions: Arena::new(),
         }
     }
 }
@@ -181,6 +185,32 @@ impl<VS: VersionSet, N: PackageName> Pool<VS, N> {
     /// Panics if the version set is not found in the pool
     pub fn resolve_version_set_package_name(&self, id: VersionSetId) -> NameId {
         self.version_sets[id].0
+    }
+
+    /// Interns a union of two or more version sets and returns its [`VersionSetUnionId`].
+    ///
+    /// Version set unions are *not* deduplicated, and a unique id is returned on every
+    /// invocation.
+    pub fn intern_version_set_union(
+        &self,
+        first: VersionSetId,
+        others: impl Iterator<Item = VersionSetId>,
+    ) -> VersionSetUnionId {
+        self.version_set_unions
+            .alloc(others.fold(SmallVec::one(first), |mut vec, version_set| {
+                vec.push(version_set);
+                vec
+            }))
+    }
+
+    /// Returns the version sets in the version set union with the given id.
+    ///
+    /// Panics if there is no union with the given id.
+    pub fn resolve_version_set_union(
+        &self,
+        id: VersionSetUnionId,
+    ) -> impl Iterator<Item = VersionSetId> + '_ {
+        self.version_set_unions[id].iter().copied()
     }
 }
 
--- a/tests/solver.rs
+++ b/tests/solver.rs
@@ -22,7 +22,7 @@ use resolvo::{
     snapshot::{DependencySnapshot, SnapshotProvider},
     utils::{Pool, Range},
     Candidates, Dependencies, DependencyProvider, Interner, KnownDependencies, NameId, SolvableId,
-    Solver, SolverCache, StringId, UnsolvableOrCancelled, VersionSetId,
+    Solver, SolverCache, StringId, UnsolvableOrCancelled, VersionSetId, VersionSetUnionId,
 };
 use tracing_test::traced_test;
 
@@ -318,6 +318,13 @@ impl Interner for BundleBoxProvider {
 
     fn solvable_name(&self, solvable: SolvableId) -> NameId {
         self.pool.resolve_solvable(solvable).name
+    }
+
+    fn version_sets_in_union(
+        &self,
+        _version_set_union: VersionSetUnionId,
+    ) -> impl Iterator<Item = VersionSetId> {
+        std::iter::empty()
     }
 }
 
--- a/tests/solver.rs
+++ b/tests/solver.rs
@@ -1,5 +1,6 @@
 use std::{
     any::Any,
+    borrow::Borrow,
     cell::{Cell, RefCell},
     collections::HashSet,
     fmt::{Debug, Display, Formatter},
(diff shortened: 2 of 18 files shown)


src/internal/mapping.rs
```diff
@@ -196,7 +196,7 @@ impl<K: ArenaId, V: serde::Serialize> serde::Serialize for Mapping<K, V> {
         self.chunks
             .iter()
             .flatten()
-            .take(self.max())
+            .take(self.max() + 1)
```
baszalmstra, line 199: Nice find! This in itself might be worth a PR!

tests/solver.rs
```diff
@@ -1105,9 +1191,11 @@ fn test_snapshot() {
         ("icons", 1, vec![]),
         ("intl", 5, vec![]),
         ("intl", 3, vec![]),
+        ("union", 1, vec!["icons 2 | intl"]),
```
baszalmstra, line 1194: Please just add a new test instead of modifying an existing one.

baszalmstra (commented):
This is awesome!

I think it could be useful to modify the `decide` function as well to be aware of multiple version sets in a single requirement. That way you get to conflicts of individual packages faster.

src/internal/mapping.rs
eviltak: If you prefer, I can cherry pick the commit onto a new branch and submit a new PR?

baszalmstra: That would be great!

src/utils/pool.rs
```diff
@@ -182,6 +186,32 @@ impl<VS: VersionSet, N: PackageName> Pool<VS, N> {
     pub fn resolve_version_set_package_name(&self, id: VersionSetId) -> NameId {
         self.version_sets[id].0
     }
+
+    /// Interns a union of two or more version sets and returns its [`VersionSetUnionId`].
+    ///
+    /// Version set unions are *not* deduplicated, and a unique id is returned on every
```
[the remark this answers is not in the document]
eviltak: Version set unions always are just resolved to the underlying version sets, so I saw no tangible benefit (apart from not storing duplicate entries in the `Mapping`) for avoiding duplicate version set unions. Unlike say version sets, no heavy operations (like querying the `DependencyProvider`) need to performed for version set unions, so I considered it to be more effort than it was worth.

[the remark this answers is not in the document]
baszalmstra: Im a bit on the fence about that. The `Pool` is a utility thing, Im not sure if we should assume aspects about the implementation. But on the other hand we can always change that later. So lets leave as it.

eviltak:
> I think it could be useful to modify the `decide` function as well to be aware of multiple version sets in a single requirement. That way you get to conflicts of individual packages faster.

@baszalmstra Can you elaborate on how this might work? As far as I understand, `decide` currently picks the first selectable candidate in an undecided clause that has the least number of undecided candidates. The candidates list contains the candidates for all the version sets in the requirement, so it should already take into account this case.

baszalmstra:
I think you only have to look at the first versionset in a requirement that still has undecided candidates. The idea is to force conflicts as early as possible to reduce backtracking. By only looking at the first versionset instead of the whole requirement you find smaller sets faster is my theory.

E.g.

foo ==1 | bar >0,<1000 

has a large number of candidates but foo==1 only has 1. We want to select foo=1 first to make sure that if there are other conflicting requirements we find them soon.

Does that make sense?

eviltak:
I think that makes sense. Would sorting the candidates list for a version set union such that the version sets with fewer candidates come earlier suffice (in `SolverCache::get_or_cache_sorted_candidates`), or would you rather `Solver::decide` look at the candidates of the smaller version sets first?

baszalmstra:
Changing the order would also change the outcome of the solution. I can imagine that the order of the union is therefore important. So I think this order should recide in the decide function itself.

baszalmstra:
you also cannot know the number of available candidates beforehand because it depends on previous decisions

eviltak:
I seemed to have misunderstood your suggestion at first. Having re-read it a few times, I now think I have understood what you are suggesting - to consider the size of only the _first_ version set with available candidates in a requirement in the "least available candidates" heuristic of `Solver::decide`. 

If my new understanding of your suggestion is correct (and please correct me if I am still wrong), I fail to see how this will reduce the amount of backtracking. In the example you gave of `foo == 1 | bar >0,<1000`, though there is only 1 candidate for `foo` that fulfills the requirement, there are 999 candidates of `bar` that will also fulfill the requirement. If selecting `foo=1` causes conflicts, there are still the 999 candidates of `bar` to be considered. In fact, it would be easier to first try and solve other clauses with fewer available candidates, in case one of them needs a version of `foo` other than 1 (i.e. conflicts with `foo=1`), so that when we finally choose to decide a solvable for this clause, `foo=1` would have already been set to `false`, letting us pick one of the `bar` candidates instead to fulfill this requirement (which is the same justification for the heuristic when requirements can only span a single version set).

For example, consider a solution which needs two requirements `foo == 1 | bar >0,<1000` and `foo >1,<5`. With the current logic, the latter requirement will be decided first, setting say `foo=4` to true. Then the former requirement will be decided, at which point `foo=1` is no longer a viable candidate, and hence a solvable for `bar` is chosen. With your proposal (assuming I understood it correctly), we would decide the _former_ requirement first and choose `foo=1` to satisfy it. However, after this choice, the propagation step would report a conflict with the latter requirement, and we would learn that `foo=1` cannot be part of the solution and have to backtrack.

baszalmstra:
> I now think I have understood what you are suggesting - to consider the size of only the first version set with available candidates in a requirement in the "least available candidates" heuristic of Solver::decide.

Indeed.

Maybe instead of theorizing about the implications, we should just benchmark it? I assume you have a large set of cases that you could test this with?

eviltak:
I don't have test cases on hand, but I can definitely come up with some. How are you currently benchmarking and comparing the performance of various solution strategies? Do you look at the number of solver iterations, the number of times it backtracks, or some other metrics? If it works for you, I think it would be better to merge this PR as is and tackle this question in a separate issue/PR.

Also, would you be able to look at the updated C++ bindings? I added support for using `Requirement`s as I saw fit, but the API is unfortunately not as ergonomic as it is in Rust because of the lack of support of sum types in C++.

baszalmstra:
> I don't have test cases on hand, but I can definitely come up with some. How are you currently benchmarking and comparing the performance of various solution strategies? Do you look at the number of solver iterations, the number of times it backtracks, or some other metrics? If it works for you, I think it would be better to merge this PR as is and tackle this question in a separate issue/PR.

You can create a snapshot of a couple of cases with your `DependencyProvider` implementation. We can include those snapshots in this repository and use them for tests and benchmarks. 

I have a very large snapshot locally of the entire conda ecosystem (using [this tool](https://github.com/conda-incubator/rattler/tree/main/tools/create-resolvo-snapshot)), I then solve for each individual package in that set (20k+) and measure the performance using [this tool included in the repo](https://github.com/mamba-org/resolvo/tree/main/tools/solve-snapshot). You can then see a nice distribution of the solve times and compare them against another implementation. 

> Also, would you be able to look at the updated C++ bindings? I added support for using Requirements as I saw fit, but the API is unfortunately not as ergonomic as it is in Rust because of the lack of support of sum types in C++.

I can take a look at that, but next week at the earliest.

Thanks again for this PR, you are doing amazing work! 🚀

eviltak:
That's great information, thanks a lot! I should be able to generate real-world use cases with union requirements as snapshots from my APT `DependencyProvider` implementation once this PR is merged. I also think it will be better to include these benchmarks as part of the suite of checks run on every commit to keep track of performance improvements and regressions, if you are open to it. I'll wait for you to review the C++ binding API considering this PR ready to merge, and tackle the benchmarking in a subsequent PR.

baszalmstra (approved):

baszalmstra:
Thanks @eviltak, also for the C++ bindings! This looks good to me!

PR #58, opened 2024-07-29
fix: Fix off-by-one error in `Mapping::serialize`

eviltak:
Fixes an off-by-one error in the implementation of `Mapping::serialize`, and adds a serialization test (which runs under the `serde` feature) as well.


baszalmstra:
Much appreciated! Thank you!

PR #62, opened 2024-08-08
feat(solver)!: Make `Problem` use builder pattern

eviltak:
Changes the `Problem` input struct (introduced in #54) to use the builder pattern.

This allows the `soft_requirements` field to be an `Iterator` rather than a `Vec`, which prevents having to unnecessarily clone the `SolvableId`s passed to `soft_requirements` (which in some cases could be on the order of hundreds of thousands). Also, this prevents adding a new input argument to `Solver::solve` (i.e. a new field in `Problem`) from being a breaking change.

The C++ bindings remain unchanged and rely on using a simple `resolvo_cpp::Problem` struct as before; they do not expose the builder pattern.

### DIFF
--- a/src/solver/mod.rs
+++ b/src/solver/mod.rs
@@ -35,17 +35,60 @@ struct AddClauseOutput {
 }
 
 /// Describes the problem that is to be solved by the solver.
-#[derive(Default)]
-pub struct Problem {
-    /// The requirements that _must_ have one candidate solvable be included in the
+///
+/// This struct is generic over the type `S` of the collection of soft requirements passed
+/// to the solver, typically expected to be a type implementing [`IntoIterator`].
+///
+/// This struct follows the builder pattern and can have its fields set by one of the available
+/// setter methods.
+pub struct Problem<S> {
+    requirements: Vec<Requirement>,
+    constraints: Vec<VersionSetId>,
+    soft_requirements: S,
+}
+
+impl Default for Problem<std::iter::Empty<SolvableId>> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Problem<std::iter::Empty<SolvableId>> {
+    /// Creates a new empty [`Problem`]. Use the setter methods to build the problem
+    /// before passing it to the solver to be solved.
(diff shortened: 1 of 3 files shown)


src/solver/mod.rs
```diff
@@ -35,17 +35,57 @@ struct AddClauseOutput {
 }
 
 /// Describes the problem that is to be solved by the solver.
-#[derive(Default)]
-pub struct Problem {
-    /// The requirements that _must_ have one candidate solvable be included in the
+///
+/// This struct follows the builder pattern and can have its fields set by one of the available
+/// setter methods.
+pub struct Problem<S> {
```
baszalmstra, line 41: Im wondering what makes soft_requirements special conpared to requirements. Why can we use a Vec for the soft requirements but not for the other fields?

eviltak: The `soft_requirements` are only iterated over once in the `Solver::solve` method, unlike `requirements` and `constraints` which need to be stored in the `root_requirements` and `root_constraints` fields of the `Solver` struct (so they can be cloned and specified in the `Dependencies` of the root solvable whenever it is requested).

eviltak: We could very well choose to use an iterator for `requirements` and `constraints` as well, but they would end up being `collect`ed into a `Vec` to be stored in the `root_requirements` and `root_constraints` fields of the `Solver` struct anyway.

src/solver/mod.rs
```diff
@@ -35,17 +35,57 @@ struct AddClauseOutput {
 }
 
 /// Describes the problem that is to be solved by the solver.
-#[derive(Default)]
-pub struct Problem {
-    /// The requirements that _must_ have one candidate solvable be included in the
+///
+/// This struct follows the builder pattern and can have its fields set by one of the available
+/// setter methods.
```
baszalmstra, line 40: Maybe just add in comments why S is generic for future readers. Then it looks good to me.

baszalmstra (approved):

baszalmstra:
Thanks as always! :)

PR #66, opened 2024-09-26
perf: visit less clauses during propagation

baszalmstra:
This branch adds several optimizations. From large impact to small they are:

* We implement two watch literals by monitoring literals now instead of solvables. This removes a lot of unnecessary work.
* It refrains from adding forbidden clauses for packages that will never be selected anyway. This reduces the number of clauses and thus improving backtracking performance. 
* We also reduced the memory size of clauses from 24 to 16 bytes per clause, improving cache locality.
* Furthermore we investigated forcibly inlining certain code paths which also showed a small performance improvement.

### DIFF
--- a/src/solver/clause.rs
+++ b/src/solver/clause.rs
@@ -2,10 +2,10 @@ use elsa::FrozenMap;
 use std::ops::ControlFlow;
 use std::{
     fmt::{Debug, Display, Formatter},
-    hash::Hash,
     iter,
 };
 
+use crate::internal::arena::ArenaId;
 use crate::{
     internal::{
         arena::Arena,
@@ -291,6 +291,13 @@ impl Clause {
             },
         );
     }
+
+    pub fn display<'i, I: Interner>(&self, interner: &'i I) -> ClauseDisplay<'i, I> {
+        ClauseDisplay {
+            kind: *self,
+            interner,
+        }
+    }
 }
 
 /// Keeps track of the literals watched by a [`Clause`] and the state associated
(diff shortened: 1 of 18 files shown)


wolfv:
Two flamegraphs, before and after. The timing difference is _huge_ - after this PR it finishes in a second or so, and before it took more than 30 secs before I killed it.

after flamegraph

before flamegraph

baszalmstra:
Here are the benchmark results compared to the main branch:

Untitled
Untitled

This is a clear improvement.

@jjerphan You might be interested in this one!

jjerphan:
Hi @baszalmstra and @wolfv,

At first sight, the performance improvements are quite significant.

I am definitely interested in having a look at this and at other ones. I need to find some proper time to review it.

On a side note, I would like to implement an ILP formulation of the problem (as first highlighted by @DerThorsten): this offers theoretical consistency but might not offer tractability in practice. I wrote [a few notes about it](https://hackmd.io/@jjerphan/dependency-resolution-as-ilp/).

src/solver/clause.rs
```diff
@@ -356,249 +379,204 @@ impl ClauseState {
...
-    pub(crate) solvable_id: InternalSolvableId,
-    pub(crate) negate: bool,
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub(crate) struct Literal(u32);
+
+impl Literal {
+    /// Constructs a new [`Literal`] from a [`InternalSolvableId`] and a boolean
+    /// indicating whether the literal should be negated.
+    pub fn new(solvable_id: InternalSolvableId, negate: bool) -> Self {
+        assert!(solvable_id.0 < (u32::MAX >> 1) - 1, "solvable id too big");
```
wolfv, line 518: Could / should this be a debug assert? I assume we're not constructing so many literals on the fly anymore, so perhaps this is fine?

src/solver/mod.rs
```diff
@@ -1078,41 +1126,43 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
     ) -> Result<u32, Conflict> {
         {
             tracing::info!(
-                "├─ Propagation conflicted: could not set {solvable} to {attempted_value}",
+                "├┬ Propagation conflicted: could not set {solvable} to {attempted_value}",
```
wolfv, line 1129: This is such beautiful ascii pipes.

src/solver/mod.rs
```diff
@@ -1195,46 +1246,56 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
...
 
                 // Get mutable access to both clauses.
-                let mut clauses = self.clauses.borrow_mut();
-                let (predecessor_clause, clause) =
+                let (predecessor_clause_state, clause_state) =
                     if let Some(prev_clause_id) = predecessor_clause_id {
-                        let (predecessor_clause, clause) =
-                            clauses.get_two_mut(prev_clause_id, clause_id);
-                        (Some(predecessor_clause), clause)
+                        let prev_idx = prev_clause_id.to_usize();
```
wolfv, line 1288: I think internally these are i32? Could we skip the usize casting and compare the inner values directly? Not important though.

PR #79, opened 2024-11-01
fix: constraint at root can conflict

wolfv:
This fixes an issue that Jaime found in his work on `conda-rattler-solver`. 

In conda-rattler-solver a lot more constraints are used then we generally do in pixi and other projects, so this became apparent quite quickly.

I haven't been able to verify whether the message makes a lot of sense yet, though.

### DIFF
--- a/src/conflict.rs
+++ b/src/conflict.rs
@@ -985,7 +985,7 @@ impl<'i, I: Interner> fmt::Display for DisplayUnsat<'i, I> {
                     &ConflictCause::Constrains(version_set_id) => {
                         writeln!(
                             f,
-                            "{indent}The constraint at the root level {version_set} could not be fulfilled",
+                            "{indent}root constraint '{version_set}' cannot be fulfilled",
                             version_set = self.interner.display_version_set(version_set_id),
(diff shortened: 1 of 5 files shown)


src/conflict.rs
```diff
@@ -982,7 +982,14 @@ impl<'i, I: Interner> fmt::Display for DisplayUnsat<'i, I> {
 
                 // The only possible conflict at the root level is a Locked conflict
                 match conflict {
-                    ConflictCause::Constrains(_) | ConflictCause::ForbidMultipleInstances => {
+                    &ConflictCause::Constrains(version_set_id) => {
+                        writeln!(
+                            f,
+                            "{indent}The constraint at the root level {version_set} could not be fulfilled",
```
baszalmstra, line 988: I think its obvious for the user that the constraint is at the root level. I would just remove that part.

wolfv: The message is now `constraint "foo" cannot be fulfilled`. I added a test with a insta snapshot.

baszalmstra (approved):

PR #85, opened 2024-11-28
chore: add diagnostics and segmented vector

baszalmstra:
This PR adds the use of `segvec::SegVec` to allocate clauses. This reduces the number of very large allocations performed by the library.

I also added functionality to report diagnostics to make it easier to debug what is going on.

### DIFF
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Run rustfmt
         uses: actions-rust-lang/rustfmt@v1
       - name: Run clippy
-        run: cargo clippy --workspace
+        run: cargo clippy --workspace --all-features
 
   test:
     name: Test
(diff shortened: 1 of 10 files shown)


wolfv:
Cool stuff! Did you benchmark that using `segvec` is beneficial?

The diagnostics are very nice.

baszalmstra:
I have not, I will run a benchmark at the end of the day.

baszalmstra:
But without it, I ran into issue with the solver trying to allocate 1.5GB in one go, which causes an OOM.

baszalmstra:
These are the timings results:

Untitled

There is no significant increase in outliers, but on average, the code became slightly slower. 

Untitled-1

Ill try removing the segvec later and try again to see if thats the issue.

tdejager:
Now prints something like:

```
Total number of solvables:  4
Total number of watches:    6 (512 B)
Total number of variables:  4 (108 B)
Total number of clauses:    5 (160 B)
- Requires:                 1
- Constrains:               0
- Lock:                     0
- Learned:                  0
- ForbidMultiple:           2
  - opentelemetry-api:      2
```
  
Removed SegVec cause it seems slower.

PR #92, opened 2025-01-01
refactor: reduce watchmap memory size

baszalmstra:
Reduce the memory consumption of the `WatchMap` by utilizing `NonZeroU32`. Rust has an optimization that an `Option<NonZero..>` is the same size as a `NonZero..`. The optimization uses the fact that `0` means `None`. 

Our `Mapping` type stores `Option<T>`. The `WatchMap` contains a `Mapping<Literal, ClauseId>`. For each `Literal` we store an `Option<ClauseId>`. Previously we would encode a `null` clause as `ClauseId::null`. We now store this as an `Option<ClauseId>`.

This reduces the memory size of an `Option<ClauseId>` from 8 bytes to 4 bytes. It reduces the memory usage of the watchmap by 50% and makes working with a null clause more ergonomic.

I didnt run the benchmarks but I am quite confident that there will be no negative impact.

### DIFF
--- a/src/internal/id.rs
+++ b/src/internal/id.rs
@@ -1,4 +1,7 @@
-use std::fmt::{Display, Formatter};
+use std::{
+    fmt::{Display, Formatter},
+    num::NonZeroU32,
+};
 
 use crate::{internal::arena::ArenaId, Interner};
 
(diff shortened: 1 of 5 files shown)


wolfv:
Nice!

wolfv (approved):

PR #101, opened 2025-01-17
feat: add support for conditional dependencies

prsabahrami:


src/solver/clause.rs
```diff
@@ -230,6 +232,45 @@ impl Clause {
...
+    fn conditional_impl(
+        package_id: VariableId,
+        condition: VariableId,
+        then: Requirement,
+        candidates: impl IntoIterator<Item = VariableId>,
+        decision_tracker: &DecisionTracker,
+    ) -> (Self, Option<[Literal; 2]>, bool) {
+        // It only makes sense to introduce a conditional clause when the package is undecided or going to be installed
+        assert_ne!(decision_tracker.assigned_value(package_id), Some(false));
+        assert_ne!(decision_tracker.assigned_value(condition), Some(false));
```
baszalmstra, line 244: The condition might already have any value assigned.

src/solver/clause.rs
```diff
@@ -230,6 +232,45 @@ impl Clause {
         )
     }
 
+    fn conditional_impl(
+        package_id: VariableId,
+        condition: VariableId,
+        then: Requirement,
+        candidates: impl IntoIterator<Item = VariableId>,
```
baszalmstra, line 239: Its unclear which candidates these are? Is it for the condition, the requirement or the combined set?

prsabahrami:
@baszalmstra In the `decide` function, I am chaining the conditional clauses with the requires clauses and iterate over both of the clauses but skip the conditionals if non of the conditions are set to true. Let me know what you think!

src/solver/clause.rs
```diff
@@ -230,6 +234,67 @@ impl Clause {
...
+                    // All conditions false
+                    (
+                        Clause::Conditional(parent_id, condition, requirement),
+                        None,
+                        false,
+                    )
+                }
+            }
+        }
+    }
```
baszalmstra, line 296: I think this code can be simplified. We need to do the following things:

We need to find two literals (either from the parent, the condition or the requirement) which are either undecided or set to true. Since the assertion at the top already indicates that the parent meets that condition you only need to find a single literal that meets that condition.

If no such literal can be found, e.g. all the literals are false, it indicates there is a conflict and you should select a random literal (just the first for example).

There is one edge case:
- If there are no condition candidates. That is an assertion. If we would otherwise add the clause it would just be a requires clause (e.g. ~A v D1 v ... D2) which is not what we want.

src/solver/clause.rs
```diff
@@ -272,6 +337,22 @@ impl Clause {
             Clause::Lock(_, s) => [s.negative(), VariableId::root().negative()]
                 .into_iter()
                 .try_fold(init, visit),
+            Clause::Conditional(package_id, condition, requirement) => {
+                iter::once(package_id.negative())
+                    .chain(
+                        requirements_to_sorted_candidates[&condition.into()]
```
baszalmstra, line 343: This doesnt need to be sorted at all. :)

tests/solver.rs
```diff
@@ -1429,6 +1432,203 @@ fn test_explicit_root_requirements() {
...
+    provider.add_package("c", 1.into(), &[], &[]); // Simple package c
+
+    // Create conditional requirement: if b=1 is installed, require c
+    let b_spec = Spec::parse_union("b 1").next().unwrap().unwrap();
+    let c_spec = Spec::parse_union("c 1").next().unwrap().unwrap();
+
+    let b_version_set = provider.intern_version_set(&b_spec);
+    let c_version_set = provider.intern_version_set(&c_spec);
+
+    let conditional_req = ConditionalRequirement::new(b_version_set, c_version_set.into());
```
baszalmstra, line 1452: Ideally you add a little string conversion for parsing this so you can do e.g. `c 1; if b 1..2` that would make the tests more readable.

tests/solver.rs
```diff
@@ -1429,6 +1432,203 @@ fn test_explicit_root_requirements() {
...
+    let result = transaction_to_string(solver.provider(), &solved);
+    // b and target should be installed
+    insta::assert_snapshot!(result, @r###"
+        b=1
+        target=1
+        "###);
+}
+
+#[test]
+fn test_circular_conditional_dependencies() {
```
baszalmstra, line 1596: I think you should add a few tests that include more versions for a single package.

baszalmstra (commented):
Partial review, will review the rest later.

src/solver/clause.rs
prsabahrami: I've added another map to `DependencyProvider` which maps `VersionSetId` to `VariableIds`. The cache did not directly provide this and I don't think we need it in the cache as we are only using it in two places and so it would make more sense to have them on the `DependencyProvider` side.

baszalmstra: I assume you mean the Solver instead of the DependencyProvider?

Just a thought but Im wondering if we ever care about the solvables instead of the variables? Perhaps it makes sense to store that in the cache instead?

wolfv:
When I look at the tests, it looks like it's hitting some asserts? :)

tests/solver.rs
```diff
@@ -113,31 +113,46 @@ impl FromStr for Pack {
...
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let split = s.split(' ').collect::<Vec<_>>();
-        let name = split
-            .first()
-            .expect("spec does not have a name")
-            .to_string();
+        let split = s.split(';').collect::<Vec<_>>(); // c 1; if b 1..2
+
+        if split.len() == 1 {
```
baszalmstra, line 141: You can also use `split_once("; if")`

tests/solver.rs
```diff
@@ -113,31 +113,46 @@ impl FromStr for Pack {
...
+            let split = s.split(' ').collect::<Vec<_>>();
+            let name = split
+                .first()
+                .expect("spec does not have a name")
+                .to_string();
+            let versions = version_range(split.get(1));
+            return Ok(Spec::new(name, versions, None));
+        }
+
+        let binding = split.get(1).unwrap().replace("if", "");
```
baszalmstra, line 152: You can use `strip_prefix`

src/solver/clause.rs
```diff
@@ -230,6 +234,63 @@ impl Clause {
...
+                Some([parent_id.negative(), req.positive()]),
+                false,
+            ),
+            // No valid literals found - conflict case
+            (None, None) => (
+                Clause::Conditional(parent_id, condition, requirement),
+                Some([parent_id.negative(), first_condition.negative()]),
+                true,
+            ),
+        }
```
baszalmstra, line 291: You can map the two iterators to their respective literals, then chain them, and eval the _literals_ to find one you need.

baszalmstra (commented):
I was thinking, there might be a bug in the code as is. 

Given the conditional requirement: 
`A requires C 1..3; if B 1..2`

e.g.

`A∧(B1∨B2)⟹C1∨C2∨C3`

in CNF form that becomes:

`¬A∨¬B1∨C1∨C2∨C3`
`¬A∨¬B2∨C1∨C2∨C3`

But currently we convert it to one clause:

`¬A∨¬B1 v ¬B2∨C1∨C2∨C3`

Which will always evaluate to `true` if any version of `B` is selected.

prsabahrami:
@baszalmstra I have added two tests with more versions. One is the example you provided above `A requires C 1..3; if B 1..2`. The second one is to confirm the behaviour of the solver when the highest version should be installed. 
`A requires B 1..3, A requires C 1..3 if B 1..2`. In this example, `C` should not be installed as the resolver picks `B 2` as the dependency of `A`.

baszalmstra:
Superseded by #136 . Have a look @prsabahrami !

PR #108, opened 2025-02-06
Avoid panic in Itertools::format_with

jrray:
Run this new test by itself with `cargo test test_tracing`.

When tracing is configured with multiple output channels, it is possible for the `tracing::debug!` macro to call `fmt` multiple times on its arguments.

`format_with` is documented to panic if formatted more than once. https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.format_with

The panic can be avoided by adding `.to_string()` to the end of the use of `format_with` around line 569 of solver/mod.rs, but this may not be the best fix.

This test sets a global default which conflicts with tracing_test, so this new test doesn't play nice with the other tests when running the whole test suite.

Edit:

Test removed and the `to_string` fix has been put in place instead.

### DIFF
--- a/src/solver/mod.rs
+++ b/src/solver/mod.rs
@@ -572,6 +572,7 @@ impl<D: DependencyProvider, RT: AsyncRuntime> Solver<D, RT> {
                         self.clauses.kinds[derived_from.to_usize()]
                             .display(&self.variable_map, self.provider()),
                     )))
+                    .to_string()
             );
             tracing::debug!("====");
 


jrray:
Oops didn't mean to reformat Cargo.toml. This PR is not intended to be merged anyway.

baszalmstra:
I had to read it a couple of times but now it makes sense to me. I would be fine adding `to_string` at the end! To me the most important thing is that we do not pay for allocate the string if the log is also not going to be visible. I think adding `to_string` behind the format_with achieves this though. 

Would you mind implementing the fix too?

jrray:
Yes, I'm happy to. I have been running with that change already in an internal build and it's been fine.

I just don't know how to implement a test for this problem that won't interfere with the other tests. 

Would you prefer I modify this PR to add the fix (and remove the test) or submit a new PR?

baszalmstra:
Awesome, feel free to modify this pr

baszalmstra (approved):

PR #131, opened 2025-04-22
refactor: encoding of clauses

baszalmstra:
This PR refactors how we get information from the dependency provider and encode the information into sat clauses. Functionally nothing changes but I refactored the `add_clauses_for_solvables` method into an object to make it much clearer what the code flow is.


baszalmstra:
Still need to do a benchmark but I doubt this will impact it at all.

tdejager (approved):
Very nice, much cleaner and seems more rusty in general! 👍

PR #136, opened 2025-05-08
feat!: add support for conditional dependencies

baszalmstra:
This PR adds support for conditional dependencies. A conditional dependency is a dependency on a package that is only active if another requirement is satisfied. E.g. `A requires B if C`. Conditions can be combined with `and` and `or`. This allows depending on packages that are only active in certain scenarios, e.g.

* Depending on the platform (using conda virtual packages)
* Depending on the presence of other packages (e.g. `numpy <=1.22; if python<3.12`, `numpy >1.22; if python >=3.13`).

Care is taken that conditional dependencies only incur an overhead when used. So this PR should not affect the performance of problems without conditional dependencies. Even when they are used the overhead should be minimal.

I've added a lot of documentation that describes how conditional dependencies are implemented.

Todo

- [ ] Benchmark
- [ ] Cpp bindings


baszalmstra:
@prsabahrami @jjerphan If you have time I would appreciate a few more eyes on this PR! 👍

src/internal/id.rs
```diff
@@ -56,6 +56,36 @@ impl ArenaId for VersionSetId {
...
 
+/// The id associated with a Condition.
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+pub struct ConditionId(NonZero<u32>);
+
+impl ConditionId {
+    /// Creates a new `ConditionId` from a `u32`, panicking if the value is zero.
```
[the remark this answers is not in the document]
baszalmstra:
```suggestion
    /// Creates a new `ConditionId` from a `u32`
```

src/solver/conditions.rs
```diff
@@ -0,0 +1,124 @@
...
+//! Note that we do not encode any clauses to force `C_selected` to be false.
+//! We argue that this state is not required to function properly. If
+//! `C_selected` would be set to false it would mean that all candidates of
+//! package C are unselectable. This would disable the requirement, e.g. B
+//! shouldnt be selected for A1. But it doesnt prevent A1 from being selected.
+//!
+//! Conditions are expressed as boolean expression trees. Internally  they are
+//! converted to Disjunctive Normal Form (DNF). A boolean expression is
+//! in DNF if it is a **disjunction (OR)** of one or more **conjunctive clauses
+//! (AND)**.
```
tdejager, line 55:
```suggestion
//! in DNF if the operands on the variables inside the clauses are **AND** while the clauses themselves are connected by **OR**'s
```

tdejager: I first interpreted the other as the other way around, this is my attempt to clarify while straying from the regular nomenclature. You decide!

tests/solver/bundle_box/mod.rs
```diff
@@ -0,0 +1,431 @@
...
+//
+// You can also use just a single number for a range like `package 0` which
+// means the range from 0..1 (excluding the end)
+//
+// Lets call the tuples of (Name, Version) a `Pack` and the tuples of (Name,
+// Ranges<u32>) a `Spec`
+//
+// We also need to create a custom provider that tells us how to sort the
+// candidates. This is unique to each packaging ecosystem. Let's call our
+// ecosystem 'BundleBox' so that how we call the provider as well.
```
tdejager, line 16: I think this description should be expanded to include conditionals :)

src/solver/conditions.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    /// The top-level condition to which this disjunction belongs.
+    pub _condition: ConditionId,
+}
+
+/// Converts from a boolean expression tree as described by `condition` to a
+/// boolean formula in disjunctive normal form (DNF). Each inner Vec represents
+/// a conjunction (AND group) and the outer Vec represents the disjunction (OR
+/// group).
+pub fn convert_conditions_to_dnf<I: Interner>(
```
tdejager, line 96: Could we make some unit tests for these :)?

tests/solver/main.rs
```diff
@@ -1427,20 +883,173 @@ fn test_explicit_root_requirements() {
...
+    bar=1
+    foo=1
+    icon=1
+    "###);
+}
+
+#[test]
+fn test_conditional_complex() {
+    let mut provider = BundleBoxProvider::from_packages(&[
+        ("foo", 1, vec!["icon; if bar and baz or menu"]),
```
tdejager, line 1011: Maybe another test with `menu` instead of `bar` and `baz` just to be sure.

tdejager (changes requested):
Phew! That was a read, but what excellent excellent work! Can't wait for this to land :) Just a few nits.

wolfv:
I added some benchmarking here: https://github.com/conda/rattler/pull/1541

src/utils/pool.rs
```diff
@@ -49,6 +52,12 @@ pub struct Pool<VS: VersionSet, N: PackageName = String> {
     /// Map from version set to the id of their interned counterpart
     version_set_to_id: FrozenCopyMap<(NameId, VS), VersionSetId, ahash::RandomState>,
 
+    /// Conditions that can be used to filter solvables.
+    conditions: Arena<ConditionId, Condition>,
+
+    /// Map from condition to its id
+    pub(crate) condition_to_id: FrozenCopyMap<Condition, ConditionId, ahash::RandomState>,
```
baszalmstra, line 59: is this pub crate needed?

wolfv:
Benchmark results:

image
image
image

src/utils/pool.rs
wolfv: I just kinda copied it from other fields, but probabbly it's not needed. Let me remove it for now.

Image

PR #187, opened 2025-11-26
perf: track active requirement parents

baszalmstra:
Instead of iterating over all `requires_clauses` in `decide()`, maintain an `IndexSet` of variables that are assigned true and have requires clauses. This reduces the iteration scope to only relevant variables. The idea is that this can make a big difference.


baszalmstra:
Not pursuing this any further.

PR #221, opened 2026-05-19
perf: encoding optimizations

baszalmstra:
This PR introduces a few incremental performance improvements that add up to roughly 2x performance improvement. There are several commits that each have been benchmarked individually on the conda-forge dataset.

| # | Commit | Median / problem | vs. previous | vs. main (cum.) |
|---|---|---:|---:|---:|
| 0 | main (baseline) | **1874 ms** | — | — |
| 1 | diagnostics counters + propagate split | **1776 ms** | 1.055× | 1.055× |
| 2 | dense `Vec` for variable origins | **1694 ms** | 1.048× | 1.106× |
| 3 | `IndexedSet`/`BitVec` clause dedup | **1760 ms** | 0.963× (noise) | 1.065× |
| 4 | binary-clause fast path in propagate | **1706 ms** | 1.031× | 1.098× |
| 5 | `forbid_seen` dedup | **993 ms** | **1.718×** ⭐ | 1.887× |
| 6 | `now_or_never` for encoding futures | **900 ms** | 1.104× | **2.082×** |

And some images from the combined stack:

image
image


wolfv (approved):

dralley:
@baszalmstra As it turns out I was looking at some similar optimizations as a result of some of my profiling, so thanks for that.

Question: `solvable_to_variable` *could* be dense, and in the event that the user guarantees they are dense, it might make sense to provide a dense layout option for the performance benefits?

Hashing and rehashing costs alone as a result of that map are along the lines of 6% of runtime in some of my profiles (that number might be out of date since I haven't retested after some tweaks and this PR)

baszalmstra:
Yeah if we can encode that assumption that would be interesting indeed!

dralley:
Optional feature perhaps?

In the RPM case, solvables are assigned sequentially and there will be one solvable per package in the repo, and every package is very likely to have at least one dependency (since distros hate static linking).  So overall I think it would be quite dense, especially in certain scenarios like a whole-distro Fedora upgrade or something.

PR #223, opened 2026-05-20
perf: minor performance optimizations

dralley:
Avoid some unnecessary allocations and de-allocations by re-using buffers and the intern pool

I haven't extensively benchmarked, but it was worth a couple of percent collectively for some of my specific cases

Likely mostly due to "Avoid String allocation on intern_package_name cache hits"

### DIFF
--- a/src/utils/pool.rs
+++ b/src/utils/pool.rs
@@ -1,4 +1,5 @@
 use std::{
+    borrow::Borrow,
     fmt::{Display, Formatter},
     hash::Hash,
 };
@@ -117,14 +118,14 @@ impl<VS: VersionSet, N: PackageName> Pool<VS, N> {
     /// [`Self::resolve_package_name`] function.
     pub fn intern_package_name<NValue>(&self, name: NValue) -> NameId
     where
-        NValue: Into<N>,
-        N: Clone,
+        NValue: Into<N> + AsRef<str>,
+        N: Clone + Borrow<str>,
     {
-        let name = name.into();
-        if let Some(id) = self.names_to_ids.get_copy(&name) {
+        if let Some(id) = self.names_to_ids.get_copy(name.as_ref()) {
             return id;
         }
 
+        let name = name.into();
         let next_id = self.package_names.alloc(name.clone());
         self.names_to_ids.insert_copy(name, next_id);
         next_id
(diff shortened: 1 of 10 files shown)


dralley:
Best viewed with "hide whitespace" turned on

src/utils/pool.rs
```diff
@@ -117,14 +118,14 @@ impl<VS: VersionSet, N: PackageName> Pool<VS, N> {
     /// [`Self::resolve_package_name`] function.
     pub fn intern_package_name<NValue>(&self, name: NValue) -> NameId
     where
-        NValue: Into<N>,
-        N: Clone,
+        NValue: Into<N> + AsRef<str>,
```
baszalmstra: Pool is generic on `N` so `AsRef<str>` breaks down as soon `N` is not `String`. I pushed a commit to work around that, let me know what you think.

baszalmstra:
I reran the benchmarks with my changes; the median and the p50 to p95 percentiles are within ~2% of main in both directions, which is well inside single-run noise. The clearest positive signal is at p99 (1.113x faster), where the allocation reductions help on the longest solves.

So happy to merge if you are happy with my changes.

baszalmstra (approved):

src/utils/pool.rs
```diff
@@ -114,18 +114,23 @@ impl<VS: VersionSet, N: PackageName> Pool<VS, N> {
...
+        NValue: IntoPackageName<N>,
+        N: Borrow<NValue::Borrowed> + Clone,
     {
-        if let Some(id) = self.names_to_ids.get_copy(name.as_ref()) {
+        if let Some(id) = self.names_to_ids.get_copy(name.as_borrowed()) {
             return id;
         }
 
-        let name = name.into();
+        let name = name.into_owned();
```
dralley, line 133: As a general question - I haven't dived into the deep internals much yet, but is this effectively storing the same strings twice, and thereby allocating twice, once in `names_to_ids` and once in `package_names`?

dralley: (that's the existing logic I mean, the change itself looks fine)

dralley:
Your commit looks fine, +1 to merge

src/utils/pool.rs
baszalmstra: Yes that's correct. If you see a way to get rid of that, that would be great. :)

PR #228, opened 2026-05-31
fix: resize name_activity for virtual provides candidates

dralley:
When a `DependencyProvider` uses virtual provides (solvable A provides capability B),` get_candidates("B")` returns solvable A whose `solvable_name()` is "A" — a different `NameId`. The name_activity vec was only resized in `on_candidates_available` for the queried name ("B"), never for the candidate's own name ("A"). During clause learning, `solvable_name()` returns the candidate's own `NameId`, causing an index-out-of-bounds panic.

Resize `name_activity` in `register_forbid_target`, which is the first place the encoder encounters a candidate's own NameId when building mutual-exclusivity constraints.

### DIFF
--- a/src/solver/encoding.rs
+++ b/src/solver/encoding.rs
@@ -631,6 +631,13 @@ impl<'a, 'cache, D: DependencyProvider> Encoder<'a, 'cache, D> {
     /// see [`Self::forbid_seen`] for why globally-deduplicating is safe.
     fn register_forbid_target(&mut self, name_id: NameId, variable: VariableId) {
         if self.forbid_seen.insert(variable) {
+            // Ensure name_activity covers this name_id. Candidates found via
+            // virtual provides may have a solvable_name() that differs from
+            // the queried package name, so on_candidates_available may never
+            // have resized for it.
+            if self.state.name_activity.len() <= name_id.to_usize() {
+                self.state.name_activity.resize(name_id.to_usize() + 1, 0.0);
+            }
             self.pending_forbid_clauses
                 .entry(name_id)
                 .or_default()


dralley:
This resolves a bounds check panic I experienced in testing `resolvo-rpm`

I'm not certain this is the best place for the check to go though.  Seems like `SolverState` ought to be managing the size of its own internal vecs.  If you have a better idea, I can change it and test.  Also let me know if I need to create some kind of test case for regression checking later.

src/solver/encoding.rs
```diff
@@ -631,6 +631,13 @@ impl<'a, 'cache, D: DependencyProvider> Encoder<'a, 'cache, D> {
     /// see [`Self::forbid_seen`] for why globally-deduplicating is safe.
     fn register_forbid_target(&mut self, name_id: NameId, variable: VariableId) {
         if self.forbid_seen.insert(variable) {
+            // Ensure name_activity covers this name_id. Candidates found via
+            // virtual provides may have a solvable_name() that differs from
```
baszalmstra, line 635: virtual provides is not something that resolvo itself provides so this is not immediately obvious.

baszalmstra (changes requested):
Can you add a regression test for this? Its not immediately clear to me how I can reproduce this. Im indeed not entirely sure if this is the right place to ensure the vector size.

dralley:
What's the method for producing a regression test?  Dump the solver state after loading everything (well, presumably a somewhat more minimized version) up?

baszalmstra:
You should be able to create a reproducer using the included "bundleboxprovider".

dralley:
As it turns out, the `IdMap` changes fixed this as a consequence, so no patch is needed.

https://github.com/prefix-dev/resolvo/pull/224/changes#diff-be5efb9b2aa1b2f674eb5dd8b2467bc3584157ff02d5c004a11322cf7257f072L1580

I'm just going to close this, then.

PR #237, opened 2026-06-12
perf: defer candidate loading for conditional requirements

baszalmstra:
## Summary

This PR makes conditional requirements lazy. Previously a conditional requirement (`pkg; if cond`) was encoded exactly like an unconditional one: its candidates were fetched, sorted, and turned into clauses at encode time, and those candidates' dependencies were fetched in turn, even when the gating condition never held at any point during the solve. The encoder now fetches only the *condition's* packages up front; the requirement's own candidates are loaded and its requires clause encoded the first time the condition actually fires.

On condition-heavy indexes most conditions never fire in a given solve, so most of that work (candidate fetching, sorting, interning, clause encoding, and the transitive dependency fan-out behind it) never happens at all.

## How it works

Eager vs lazy encoding of a conditional requirement

A deferred entry is checked against the current assignment on every solver iteration. The first time its condition holds, it is drained and encoded; the encoded clause carries the condition literals, so it remains globally valid and never needs re-encoding, no matter how the solver backtracks afterwards:

Lifecycle of a deferred conditional requirement

## Key Changes

- Deferred conditional requirements live in a `deferred_requirements` map keyed by `ConditionId`, one entry per DNF disjunct. At the top of every `run_sat` iteration the solver checks which deferred conditions currently hold; fired entries are drained and encoded exactly as the eager path would have (the requires clause carries the condition literals, so the clause remains valid if the firing assignment is later backtracked). An entry is encoded at most once.
- Clause construction now handles parents that are already decided false. With deferral, clauses can be encoded mid-solve for a solvable that was propagated false earlier (e.g. a candidate of a deferred requirement that another package's constraint forbade before the condition fired). `WatchedLiterals::requires`/`constrains` previously asserted this never happens; they now build the clause with a correctly computed conflict flag (a false parent satisfies the clause, so it never conflicts, and the ordinary watch machinery takes over if the assignment is backtracked). Skipping the clause instead would be unsound: nothing ever revisits a drained deferred entry or an already-encoded solvable.
- `SnapshotProvider` no longer shadows the snapshot's highest-index version set with the additional version sets created by `add_package_requirement` (this made any solve touching that version set panic or resolve the wrong set).
- `tools/solve-snapshot` gains `--skip N`: skipped problems are still generated so the RNG advances deterministically, which makes `--seed S --skip K -n K+1` replay exactly problem K of a seeded run, e.g. to re-measure an outlier in isolation.

## Performance Results

Benchmarked on the conda-pypi + conda-forge snapshot (PyPI extras modelled as conditional requirements, so the workload is condition-heavy), 1000 randomized problems per seed, two seeds, identical problem sequences on both sides, 60 s timeout. Baseline is main plus the `SnapshotProvider` fix from this branch, so the harness fix does not pollute the comparison. The screening runs were core-pinned and ran in parallel; every flagged outlier was afterwards re-measured strictly sequentially on an idle machine and reproduced its timing.

| Speedup (2000 problems) | |
|--------|---------:|
| mean | **12.9×** |
| median | 141× |
| p90 | 9.4× |
| p95 | 7.3× |
| p99 | **4.7×** |
| timeouts | 1 → 1 |

| metric | main | this PR |
|---|---:|---:|
| total solve time (seed 0) | 2071.3 s | **132.3 s (−93.6%)** |
| total solve time (seed 1) | 2178.4 s | **197.3 s (−90.9%)** |
| p50 (pooled) | 3.242 s | 0.023 s |
| p90 (pooled) | 3.368 s | 0.358 s |
| p99 (pooled) | 3.961 s | 0.846 s |
| ok / unsat / timeout | 161/839/0 and 141/858/1 | identical |

Solve duration histogram

Per-problem duration difference

Top 5 most-improved problems (this PR vs main):

- `asttokens[test], mypy-boto3[sns-with-docs], ipykernel` | **10.27s faster**
- `libgdal ==3.6.4 hada8d5e_2, mypy-boto3[qldb-with-docs], ...` | **7.76s faster**
- `mypy-boto3-mediastore-with-docs, mypy-boto3-neptune, ...` | **7.65s faster**
- `opentelemetry-exporter-otlp-proto-grpc[test], apache-airflow...` | **4.01s faster**
- `geotiff, ipaddr` | **3.93s faster**

There are no regressions: not one of the 2000 problems is both ≥1.5× and ≥50 ms slower, and the worst absolute slowdown over all 2000 problems is **2.2 ms** (on a 6 ms problem). The slowest solves on this branch (a ~2% tail between 0.75 s and 5.4 s) were each replayed in isolation with `--skip`: every timing reproduced, and main is 2.0-4.3× slower on those same problems. The single mutual 60 s timeout (seed 1, problem 480) flips in this branch's favor with a larger budget: this PR solves it in 196 s, main is still cancelled at 300 s.

## Correctness

- Statuses are identical on all 2000 problems (including the one timeout), and every solved problem returns the same number of records on both sides. The isolated replays match record counts as well.
- The 1000-problem depth is what surfaced the parent-already-false case: both seeds independently panicked around problem 905 in `WatchedLiterals::constrains` when a deferred condition fired after a constraint had forbidden one of the requirement's candidates. The fix is covered by a regression test that drives the exact scenario (a constraint propagates a gated candidate false before the condition fires, then the prefetch encodes that candidate's requires and constrains clauses) through a new `hint_dependencies_available` toggle on `BundleBoxProvider`; the test panics on the unfixed encoder.
- An earlier revision *skipped* requires clauses for false parents instead of encoding them. That is unsound across backtracking: the false assignment can be undone, and the skipped clause is never revisited, silently dropping a dependency edge. Both paths now always encode.
- Targeted tests pin the deferral semantics: encode-on-first-fire (and never twice, asserted via clause counts against a fresh re-solve), entries that never fire stay deferred and unfetched, drain order determinism, AND/OR condition firing, and survival of backtracking through a fired condition.

### DIFF
--- a/src/snapshot.rs
+++ b/src/snapshot.rs
@@ -398,7 +398,11 @@ impl<'s> SnapshotProvider<'s> {
     /// Adds another requirement that matches any version of a package.
     /// If you use "*" as the matcher, it will match any version of the package.
     pub fn add_package_requirement(&mut self, name: NameId, matcher: &str) -> VersionSetId {
-        let id = self.snapshot.version_sets.max() + self.additional_version_sets.len();
+        // Additional version sets are allocated directly after the highest
+        // index present in the snapshot. `Mapping::max()` returns the highest
+        // inserted *index* (not a count), so the first free index is one past
+        // it.
+        let id = self.snapshot.version_sets.max() + 1 + self.additional_version_sets.len();
         let package = self.package(name);
 
         let matching_candidates = package
@@ -441,8 +445,12 @@ impl<'s> SnapshotProvider<'s> {
     fn version_set(&self, version_set: VersionSetId) -> &VersionSet {
         let idx = version_set.to_index();
         let max_idx = self.snapshot.version_sets.max();
-        if idx >= max_idx {
-            &self.additional_version_sets[idx - max_idx]
+        // Indices up to and including `max_idx` belong to the snapshot;
+        // anything beyond was allocated by `add_package_requirement`. Using
+        // `>=` here would shadow the snapshot's highest-index version set
+        // with the first additional one (and panic when none was added).
+        if idx > max_idx {
+            &self.additional_version_sets[idx - max_idx - 1]
         } else {
             self.snapshot
                 .version_sets
--- a/src/solver/clause.rs
+++ b/src/solver/clause.rs
@@ -170,9 +170,12 @@ impl<N: SolverId> Clause<N> {
         condition: Option<(DisjunctionId, &[Literal])>,
         decision_tracker: &DecisionTracker,
     ) -> (Self, Option<[Literal; 2]>, bool) {
-        // It only makes sense to introduce a requires clause when the parent solvable
-        // is undecided or going to be installed
-        assert_ne!(decision_tracker.assigned_value(parent), Some(false));
+        // On the lazy-conditional path the parent may already have been
+        // propagated false by the time this clause is built (e.g. a deferred
+        // requirement whose condition fires after the parent was forbidden).
+        // The clause is currently satisfied by `¬parent`, but that assignment
+        // can be backtracked later, so the clause must still be added.
+        let parent_is_false = decision_tracker.assigned_value(parent) == Some(false);
 
         let kind = Clause::Requires(parent, condition.map(|d| d.0), requirement);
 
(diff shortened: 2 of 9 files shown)


jjerphan:
🎉

dralley:
Decent chance this helps with RPM a fair amount, or will help, once I get "rich dependencies" supported.

PR #239, opened 2026-06-12
feat: support multiversion packages

dralley:
Add `allow_multiple: bool` to `Candidates`, allowing dependency providers to indicate that multiple versions of a package may coexist in the solution (e.g. kernel packages on RPM systems). When set, the solver skips the at-most-one constraint encoding for that package name.

(multiple co-existing kernels on a system is a common configuration because GRUB can be configured to present previous versions of the kernel as emergency backup boot options)

### DIFF
--- a/src/solver/encoding.rs
+++ b/src/solver/encoding.rs
@@ -331,6 +331,10 @@ impl<'a, 'cache, D: DependencyProvider> Encoder<'a, 'cache, D> {
             self.cache.provider().display_name(name_id)
         );
 
+        if package_candidates.allow_multiple {
+            self.state.allow_multiple_names.insert(name_id);
+        }
+
         // If there is a locked solvable, forbid all other candidates
         if let Some(locked_solvable_id) = package_candidates.locked {
             self.add_locked_package_clauses(locked_solvable_id, &package_candidates.candidates);
@@ -1070,7 +1074,11 @@ impl<'a, 'cache, D: DependencyProvider> Encoder<'a, 'cache, D> {
     /// Record `variable` as a candidate for a forbid-multiple clause under
     /// `name_id`. Returns silently if the variable has already been registered;
     /// see [`Self::forbid_seen`] for why globally-deduplicating is safe.
+    /// Also skips registration for packages that allow multiple versions.
     fn register_forbid_target(&mut self, name_id: D::NameId, variable: VariableId) {
+        if self.state.allow_multiple_names.contains(name_id) {
+            return;
+        }
         if self.forbid_seen.insert(variable) {
             self.pending_forbid_clauses
                 .entry(name_id)
--- a/tests/solver/main.rs
+++ b/tests/solver/main.rs
@@ -2379,3 +2379,143 @@ fn shared_requires_gate_survives_backtrack() {
         }
     }
 }
+
+mod allow_multiple_versions {
+    use super::*;
+
+    /// Non-overlapping version requirements are unsolvable without `allow_multiple`.
+    #[test]
+    fn test_non_overlapping_requirements_unsat_without_allow_multiple() {
+        let mut provider = BundleBoxProvider::new();
+        provider.add_package("kernel", 1.into(), &[], &[]);
+        provider.add_package("kernel", 2.into(), &[], &[]);
+        provider.add_package("kernel", 3.into(), &[], &[]);
+        provider.add_package("kmod-a", 1.into(), &["kernel 1..2"], &[]);
+        provider.add_package("kmod-b", 1.into(), &["kernel 3..4"], &[]);
+
+        let result = solve_unsat(provider, &["kmod-a", "kmod-b"]);
+        assert_snapshot!(result);
+    }
+
+    /// When `allow_multiple` is set, non-overlapping version requirements
+    /// from different dependents are solved by installing multiple versions.
+    #[test]
+    fn test_non_overlapping_requirements() {
+        let mut provider = BundleBoxProvider::new();
+        provider.add_package("kernel", 1.into(), &[], &[]);
+        provider.add_package("kernel", 2.into(), &[], &[]);
+        provider.add_package("kernel", 3.into(), &[], &[]);
+        provider.set_allow_multiple("kernel");
+        provider.add_package("kmod-a", 1.into(), &["kernel 1..2"], &[]);
+        provider.add_package("kmod-b", 1.into(), &["kernel 3..4"], &[]);
+
+        let requirements = provider.requirements(&["kmod-a", "kmod-b"]);
+        let mut solver = Solver::new(provider);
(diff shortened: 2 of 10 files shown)


dralley:
One other example (tbh the only other example I could find) that uses multiversion is `gpg-pubkey`, a virtual package whose "version" is derived from the fingerprint of a public key trusted by the system package manager.

A different version of `gpg-pubkey` is created for each trusted public key, and all versions of `gpg-pubkey` are considered installed simultaneously.

So it's niche but thankfully a simple and unobtrusive feature to implement otherwise.

src/solver/encoding.rs
```diff
@@ -694,7 +698,11 @@ impl<'a, 'cache, D: DependencyProvider> Encoder<'a, 'cache, D> {
     /// Record `variable` as a candidate for a forbid-multiple clause under
     /// `name_id`. Returns silently if the variable has already been registered;
     /// see [`Self::forbid_seen`] for why globally-deduplicating is safe.
+    /// Also skips registration for packages that allow multiple versions.
     fn register_forbid_target(&mut self, name_id: D::NameId, variable: VariableId) {
+        if self.state.allow_multiple_names.contains(name_id) {
```
dralley, line 1079: It likely doesn't matter much but you could probably give the branch predictor a hint here that this path is unlikely to be taken.

tests/solver/main.rs
```diff
@@ -1490,3 +1490,117 @@ fn test_constrains_multiple_parents() {
...
+    provider.add_package("driver", 1.into(), &["kernel 1..100"], &[]);
+
+    let k1 = provider.solvable_id("kernel", 1u32);
+    let k2 = provider.solvable_id("kernel", 2u32);
+    let k3 = provider.solvable_id("kernel", 3u32);
+    let requirements = provider.requirements(&["app"]);
+    let mut solver = Solver::new(provider);
+    let problem = Problem::new()
+        .requirements(requirements)
+        .soft_requirements(vec![k1, k2, k3]);
```
baszalmstra: These soft_requirements construe the result dont they? Or maybe I misunderstand what this tests?

tests/solver/main.rs
```diff
@@ -1490,3 +1490,117 @@ fn test_constrains_multiple_parents() {
...
+    provider.add_package("app", 1.into(), &["kernel 2..100"], &[]);
+
+    let k1 = provider.solvable_id("kernel", 1u32);
+    let k2 = provider.solvable_id("kernel", 2u32);
+    let k3 = provider.solvable_id("kernel", 3u32);
+    let requirements = provider.requirements(&["app"]);
+    let mut solver = Solver::new(provider);
+    let problem = Problem::new()
+        .requirements(requirements)
+        .soft_requirements(vec![k1, k2, k3]);
```
baszalmstra: Im afraid I again dont understand this test.

tests/solver/main.rs
dralley: I might also be misunderstanding you, but I think there are 2 reasons which may or may not be reasonable:

1) There was a previous attempted implementation where I was running into problems because some state was stored on the encoder and I didn't realize it wasn't persistent (`Encoder` gets recreated), and the existing unit tests were actually passing anyway -- until I was doing some additional manual testing and ran into problems.  I believe this is one of the tests that was added partly because it would have caught the issues with the original implementation (which passed tests at the time, I only discovered the issue later).

2) Nobody actually *requires* multiple versions of a package, the idea is just that they are allowed to co-exist within a solution, usually because they were already installed.  The solver has no reason to include more than one version at a time in the solution unless suggested otherwise.  So the `soft_requirements` suggest otherwise.  But maybe there's a better way to do that while still fulfilling (1)

baszalmstra: > Nobody actually requires multiple versions of a package

But isn't that exactly something that is supported by this implementation? You can have `a requires b 1..2` and `c requires b 4..5`. If `b` is multiversion I would only expect this to resolve with multiversion. And also if there is an overlap in the requirements, ideally we resolve that to a common version.

dralley: That's probably a better construction, at least for this specific test, I will change it later

Do you think it's warranted to add tests or documentation around what state can and cannot be stored on the the `Encoder`?

dralley:
OK, these tests ought to be an improvement.

tests/solver/main.rs
```diff
@@ -2216,3 +2216,145 @@ fn test_decide_queue_union_duplicate_name() {
...
+    }
+
+    /// When dependents' version requirements overlap, each requirement independently
+    /// selects its best (highest) candidate. Here kmod-a picks kernel=2 (highest in 1..3)
+    /// and kmod-b picks kernel=3 (highest in 2..4), so both are installed even though
+    /// kernel=2 alone would satisfy both requirements.
+    ///
+    /// This is actually not ideal, but resolving overlapping ranges to a single common
+    /// version would require additional optimization in the solver (e.g. soft forbid clauses
+    /// or a post-solve minimization pass).
```
dralley: Should I stick to the 70 / 80-width comments?  It felt a bit cramped, but that's seemingly the convention in this file.

tests/solver/main.rs
```diff
@@ -2216,3 +2216,143 @@ fn test_decide_queue_union_duplicate_name() {
...
+    }
+
+    /// When dependents' version requirements overlap, each requirement independently
+    /// selects its best (highest) candidate. Here kmod-a picks kernel=2 (highest in 1..3)
+    /// and kmod-b picks kernel=3 (highest in 2..4), so both are installed even though
+    /// kernel=2 alone would satisfy both requirements.
+    ///
+    /// This is actually NOT ideal - installing only kernel=2 would be optimal - but resolving
+    /// overlapping ranges to a single common version would require additional optimization in
+    /// the solver (e.g. soft forbid clauses or a post-solve minimization pass).
```
dralley, line 2432: Notable.  I'm not particularly fussed about this at the moment but it is probably worth doing eventually.

dralley: Unless you want me to do it now.

dralley:
@baszalmstra Rebased + rewrote the tests

tests/solver/main.rs
dralley: @baszalmstra Let me know your thoughts.

dralley: @baszalmstra Can I at least get feedback on whether I need to deal with this now or if it can be deferred until later?  I'd like to get this PR out of the way so that I can rebase the other work

baszalmstra: Ah sorry, this slipped of my radar. I think we can deal with the optimal solution later.

baszalmstra (approved):

dralley:
Thanks!

baszalmstra:
Thank you for this!

